PR #446, opened 2017-05-23
fix left's users logout

njqaaa:
https://github.com/jumpserver/jumpserver/issues/443


njqaaa:
470fe68 is  issues 445

PR #717, opened 2017-09-20
Fix importing csv bugs

choldrim:
1. auto detect the encoding of csv
2. if id in csv is empty, let it equal 0
3. if hostname exists, give up this asset


ibuler:
或许 requirements也需要改了

PR #4342, opened 2020-07-16
fix(radius): 修复radius认证失败问题

BaiJiangJie:


dxlsunshine:
(20) Received Access-Request Id 251 from 172.16.0.2:41577 to 172.24.0.5:1812 length 63
(20)   User-Name = "username"
(20)   User-Password = "password"
(20)   NAS-Identifier = "django-radius"
(20) # Executing section authorize from file /opt/etc/raddb/sites-enabled/default

freeradius log:  username and password is not correct. input any username or password , freeradius always receive username" and  "password"

Issue #8287, opened 2022-05-23
[Bug] SAML2登录 当RelayState为空时报错


Comver:
# 错误内容
OneLogin_Saml2_Error at /core/auth/saml2/callback/
Redirect to invalid URL: 
Request Method:	POST
Request URL:	http://*******/core/auth/saml2/callback/
Django Version:	3.1.14
Exception Type:	OneLogin_Saml2_Error
Exception Value:	
Redirect to invalid URL: 
Exception Location:	/usr/local/lib/python3.8/site-packages/onelogin/saml2/utils.py, line 215, in redirect
Python Executable:	/usr/local/bin/python
Python Version:	3.8.12
Python Path:	
['/opt/jumpserver/apps',
 '/opt/jumpserver',
 '/opt/jumpserver/apps',
 '/usr/local/bin',
 '/usr/local/lib/python38.zip',
 '/usr/local/lib/python3.8',
 '/usr/local/lib/python3.8/lib-dynload',
 '/usr/local/lib/python3.8/site-packages']

# POST data
SAMLResponse | '******=='
-- | --
RelayState | ''

# 涉及代码片段
https://github.com/jumpserver/jumpserver/blob/26cf64ad2d9c66b8f492ec2929500304c38ed73f/apps/authentication/backends/saml2/views.py#L274

ibuler:
根据这个文档配置 https://docs.jumpserver.org/zh/master/admin-guide/authentication/saml2/

我们仅在 keyclark saml2 中测试

QuentinM-Hilbtec:
Same problem for us @ibuler using Okta. By default, the RelayState is blank. Leave empty breaks, setting the root URL of Jumpserver (`https://.../`) breaks, and setting just `/` (which is your [default](https://github.com/jumpserver/jumpserver/blob/26cf64ad2d9c66b8f492ec2929500304c38ed73f/apps/authentication/backends/saml2/views.py#L274, and probably what is used when RelayState is blank) breaks too.

PR #8870, opened 2022-09-15
Fix issue #8287 with blank SAML's RelayState

QuentinM-Hilbtec:

Fixes issue #8287

#### Summary of your change

Dumbly fixes redirect an empty URLs when the IDP sends blank RelayStates.
This fix was verified on two different Jumpservers.

### DIFF
--- a/apps/authentication/backends/saml2/views.py
+++ b/apps/authentication/backends/saml2/views.py
@@ -271,7 +271,10 @@ class Saml2AuthCallbackView(View, PrepareRequestMixin):
             auth.login(self.request, user)
 
         logger.debug(log_prompt.format('Redirect'))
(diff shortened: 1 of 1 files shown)


apps/authentication/backends/saml2/views.py
```diff
@@ -271,7 +271,10 @@ def post(self, request):
             auth.login(self.request, user)
 
         logger.debug(log_prompt.format('Redirect'))
-        next_url = saml_instance.redirect_to(post_data.get('RelayState', '/'))
+        redir = post_data.get('RelayState', '/')
```
ibuler, line 274: redir = post_data.get('RelayState')

Default values are not necessary and can be misleading

ibuler (approved):

PR #9275, opened 2023-01-05
Support signing SAML Request

VinoXuanyu:
Support sign SAML Request(which is turned off now and can't be turned on)

#### Summary of your change
1. signing turned on by setting 
         'authnRequestsSigned': true 
   in advanced settings on web console

2. signAlgorithm is default to 'http://www.w3.org/2000/09/xmldsig#rsa-sha1', can be specified by setting 
        'signatureAlgorithm': {your_algo} 
  in advanced settings on web console

3. if signing turned on, sp cert and key must be uploaded first on web console

- [Y] Made sure tests are passing and test coverage is added if needed.
- [Y] Made sure commit message follow the rule of [Conventional Commits specification](https://www.conventionalcommits.org/).
- [Y] Considered the docs impact and opened a new docs issue or PR with docs changes if needed.

### DIFF
--- a/apps/authentication/backends/saml2/views.py
+++ b/apps/authentication/backends/saml2/views.py
@@ -117,10 +117,15 @@ class PrepareRequestMixin:
 
         security_default = {
             'wantAttributeStatement': False,
-            'allowRepeatAttributeName': True
+            'allowRepeatAttributeName': True,
         }
         security = other_settings.get('security', {})
         security_default.update(security)
+
+        # set a default signing algo if authnRequestSigned is set to true but no sign algo provided
+        if security_default.get('authnRequestsSigned'):
+            if not security_default.get('signatureAlgorithm'):
+                security_default['signatureAlgorithm'] = 'http://www.w3.org/2000/09/xmldsig#rsa-sha1'
 
         default = {
             "organization": {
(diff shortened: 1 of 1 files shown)


apps/authentication/backends/saml2/views.py
```diff
@@ -146,12 +151,21 @@ def get_sp_settings(self):
...
-                }
+                },
             }
         }
         sp_settings['sp'].update(attrs)
         advanced_settings = self.get_advanced_settings()
         sp_settings.update(advanced_settings)
+
+        # load sp cert and key for encryption if advanced_settings.security.authnRequestsSigned is True
+        if advanced_settings['authnRequestsSigned']:
```
Hi-JWD, line 162: Is the parameter authnRequestsSigned more appropriate in security

VinoXuanyu: ops, using `get` method here would increase robustness, will update this PR.

Hi-JWD: Should the parameter format be like this: advanced_settings = {' security ': {' authnRequestsSigned' : True}}

VinoXuanyu: 1. Yap, the key point here is to set  appropriate values for  `x509cert` and `privakeyKey` in sp_settings['sp']  which is needed for python-saml library to sign the SAML request,  which is now not supported since the certs and keys are simply not loaded and hence will trigger an `can not load key` error.

2. The `authnRequestsSigned` field  is used by  `python-saml` library to tell if to sign the SAML request or not.

3.  It is recommended to add an prompt on web console to set the values for `authnRequestsSigned` for IdPs may require signature provided, and when not, the SAML Response would be `Signature Required` and lead to unsuccessful SSO Login.

O-Jiangweidong:
JumpServer will print the cert information to SP metadata regardless of whether the authnRequestsSigned parameter is enabled. If you have any questions, please describe your IDP platform environment. We will actively cooperate with testing to deal with compatibility issues

PR #11048, opened 2023-07-22
fix a latent bug when field_type belongs to int, bool and list.

huiserwang:
This PR aims at solving a latent bug when `filed_type` belongs to `int`, `bool` and `list`. This bug will cause errors when someone choose to deploy customized/official jumpserver from code.

#### Summary of your change
The `serializers.IntegerField`, `serializers.ListField`, and `serializers.BooleanField` don't pop the argument named `allow_blank` out and the `Field` class they inherit don't accept such an argument, so that this argument should be pop out when `field_type` is `int`, `bool`, `list`. Thus, I add two lines of codes to pop `allow_blank` when `field_type` exists in `data` and belongs to the type of `int`, `bool`, `list`, to avoid this error.
Details of `Field`, `IntegerField`, `ListField`, `BooleanField` can be refer to https://github.com/encode/django-rest-framework/blob/3.13.1/rest_framework/fields.py

### DIFF
--- a/apps/common/serializers/dynamic.py
+++ b/apps/common/serializers/dynamic.py
@@ -62,6 +62,8 @@ def create_serializer_class(serializer_name, fields_info):
             data['required'] = False
         data = set_default_by_type(field_type, data, field_info)
         data = set_default_if_need(data, i)
+        if field_type in ['int', 'bool', 'list'] and "allow_blank" in data.keys():
+            data.pop('allow_blank')
         field_name = data.pop('name')
         field_class = type_field_map.get(field_type, serializers.CharField)
         serializer_fields[field_name] = field_class(**data)


ibuler:
Maybe you are right，last change fix default blank value ，but required，and not has required setting 

How do you found this issue? It's so indiscoverable

huiserwang:
> Maybe you are right，last change fix default blank value ，but required，and not has required setting
> 
> How do you found this issue? It's so indiscoverable

I try to deploy jumpserver at the cluster of our lab with some modification: make `push_account` more robust when there are more customized groups should be granted for users. Thus, I modify the type of `groups` in apps/accounts/automations/push_account/host/posix/manifest.yml to 'list', and then an error occurred.

ibuler (approved):

PR #14303, opened 2024-10-15
fix: fixed the issue that websocket cannot use "Authorization" header

csi0n:
When using the ```ws://jumpserver.com/ws/ops/tasks/log/``` api, if an Authorization Header is passed into the api, the websocket connection will fail.

An error will happen when calling the authentication method in SignatureAuthentication, the authentication method in AccessTokenAuthentication will not be executed, and ultimately the authentication will fail.


BaiJiangJie (approved):

BaiJiangJie:
Thanks for the contribution. @csi0n

ibuler (approved):

PR #14406, opened 2024-11-05
feat: azure key vault

fit2bot:
feat: azure key vault<br>perf: add azure-keyvault-secrets<br>perf：azure kv api<br>perf: Translate


apps/accounts/backends/azure/entries.py
```diff
@@ -0,0 +1,70 @@
...
+        return path
+
+
+def build_entry(instance) -> BaseEntry:
+    class_name = instance.__class__.__name__
+    entry_class_name = f'{class_name}Entry'
+    entry_class = getattr(current_module, entry_class_name, None)
+    if not entry_class:
+        raise Exception(f'Entry class {entry_class_name} is not found')
+    return entry_class(instance)
```
fit2cloudrd, line 70: Based on my current knowledge cut-off of 2021-09-01 and the code you provided, there don't seem to be any apparent issues with the code itself or its functionality. However, as per your request for suggestions:
- You might consider using absolute imports if they're used more often across modules (i.e., `assets-account-templates-historical`).
- Ensure that all methods, properties, etc. follow the same naming convention.
- Optionally, use docstrings or comments within method names to clarify their purpose.

Feel free to let me know when you need updates based on your new context!

apps/accounts/backends/azure/service.py
```diff
@@ -0,0 +1,68 @@
...
+        try:
+            self.client.begin_delete_secret(name)
+        except (ResourceNotFoundError, ClientAuthenticationError) as e:
+            logger.error(f'delete: {name} {str(e)}')
+
+    def update_metadata(self, name, metadata: dict):
+        try:
+            self.client.update_secret_properties(name, tags=metadata)
+        except (ResourceNotFoundError, ClientAuthenticationError) as e:
+            logger.error(f'update_metadata: {name} {str(e)}')
```
fit2cloudrd, line 68: As an AI language model trained by OpenAI, I cannot directly examine code files. However, based on available information, there appears to be one issue that needs improvement:

There's a typo in the documentation string `if ('azure.net' in vault_url)`
It should read 
`if ('azureregional.net' in vault_url)`

Other than this correction:
The existing content of the docstring seems complete and consistent with Python conventions; it describes how to interact with Azure Key Vault via its Python SDKs.

However, you should consider adding some additional comments to explain more precisely what each method does:

```
# Initialize the AKESSVOTC CLIENT class
class AzKessVClient(AzureKeyVaultClient):

## Methods: Create, Update, Delete, Retrieve, etc...

def __init__(self,
              key_vault_url : str,
              tenant_id : str,
              client_id : str,
              client_seceret : str,
):
    super().__init__()
    # Your initialization here...
``` 

This will make it easier for others to understand and test your code!

apps/accounts/backends/azure/main.py
```diff
@@ -0,0 +1,53 @@
...
+    def _clean_db_secret(self, instance):
+        instance.is_sync_metadata = False
+        instance.mark_secret_save_to_vault()
+
+    def _save_metadata(self, instance, metadata):
+        try:
+            entry = build_entry(instance)
+            self.client.update_metadata(name=entry.full_path, metadata=metadata)
+        except Exception as e:
+            logger.error(f'save metadata error: {e}')
```
fit2cloudrd, line 53: The code is from 2021 to before that.

I see some issues:

- The `__all__` declaration was added but it seems not used correctly. It should include only public methods, parameters and variables which can be accessed at runtime using ".*". 
- A new method `_clean_db_secret` has been added with no clear purpose or description inside the docstring.

However these issues do not have much impact on overall program functionality, they're just coding best practices concerns.

PR #14498, opened 2024-11-20
feat: add license edition check

fit2bot:
feat: add license edition check


apps/authentication/mfa/face.py
```diff
@@ -32,7 +32,9 @@ def is_active(self):
 
     @staticmethod
     def global_enabled():
-        return settings.XPACK_LICENSE_IS_VALID and settings.FACE_RECOGNITION_ENABLED
+        return settings.XPACK_LICENSE_IS_VALID \
+            and settings.XPACK_LICENSE_EDITION == 'ultimate' \
+            and settings.FACE_RECOGNITION_ENABLED
 
     def get_enable_url(self) -> str:
         return reverse('authentication:user-face-enable')
```
fit2cloudrd, line 40: To ensure proper formatting of the output and enhance readability, it's recommended to maintain consistent indentation levels when writing comments and documentation strings. Here's how you might reformat the code:

```python
def is_active(self):

     # Check XPACK license validity and facial recognition setting

@staticmethod
def global_enabled():
    return (
        settings.XPACK_LICENSE_IS_VALID
        and settings.FACE_RECOGNITION_ENABLED
    )

def get_enable_url() -> str:
    
    url = reverse('authentication:user-face-enable')

class UserFaceEnableView(View):
    ...
```
Changes made include:
- Ensuring consistent spacing around operators (e.g., `settings.` instead of just `,`)
- Adding space before opening parenthesis in conditional statement (`and`, not `or`)

Optimization suggestions could be improving method names if they seem overly verbose, but the context doesn't provide information about which methods should be optimized.

For instance, consider replacing multiple checks in `global_enabled()` with an expression that allows for short-circuit evaluation, such as `lambda: ...`.

apps/jumpserver/settings/_xpack.py
```diff
@@ -18,6 +18,7 @@
 XPACK_TEMPLATES_DIR = []
 XPACK_CONTEXT_PROCESSOR = []
 XPACK_LICENSE_IS_VALID = False
+XPACK_LICENSE_EDITION = ""
 XPACK_LICENSE_INFO = {
     'corporation': corporation,
 }
```
fit2cloudrd, line 24: There are no issues found with the provided code snippet at this moment. It seems to be syntactically correct and is in line with current best practices.

The `license_is_valid` flag has a typo ('valid' instead of 'edition'), but it still means effectively that it's not valid or set yet to determine whether further actions should proceed according to license conditions.

I recommend reviewing other parts of your codebase for additional discrepancies before concluding anything about possible problems; however, given its current state:

* Potential issue: Typo in variable name (`version` -> `licens...`). Ensure proper casing throughout.
  
* Optimization suggesttion:
   - You could add imports if needed (e.g., `from django.contrib.auth import authenticate`, etc.), assuming you'll need them elsewhere.

Let me know how I can assist next!

PR #14709, opened 2024-12-20
fix: set default ldap user dn cache time (0)

fit2bot:
fix: set default ldap user dn cache time (0)


apps/jumpserver/conf.py
```diff
@@ -310,7 +310,7 @@ class Config(dict):
         'AUTH_LDAP_HA_START_TLS': False,
         'AUTH_LDAP_HA_USER_ATTR_MAP': {"username": "cn", "name": "sn", "email": "mail"},
         'AUTH_LDAP_HA_CONNECT_TIMEOUT': 10,
-        'AUTH_LDAP_HA_CACHE_TIMEOUT': 3600 * 24 * 30,
+        'AUTH_LDAP_HA_CACHE_TIMEOUT': 0,
         'AUTH_LDAP_HA_SEARCH_PAGED_SIZE': 1000,
         'AUTH_LDAP_HA_SYNC_IS_PERIODIC': False,
         'AUTH_LDAP_HA_SYNC_INTERVAL': None,
```
BaiJiangJie, line 316: There were no identified differences between the two configurations provided, nor apparent issues or irregularities to optimize. Both configurations look correct based on the available information and knowledge at hand.

However, if you would like a more detailed comparison, please share specific requirements or context that might not be included in either of these examples, and I can elaborate further upon their applicability or any areas where they diverge from common standard practices.

apps/settings/serializers/auth/ldap.py
```diff
@@ -85,7 +85,7 @@ class LDAPSettingSerializer(serializers.Serializer):
     )
     AUTH_LDAP_CACHE_TIMEOUT = serializers.IntegerField(
         min_value=0, max_value=3600 * 24 * 30 * 12,
-        default=3600 * 24 * 30,
+        default=0,
         required=False, label=_('User DN cache timeout (s)'),
         help_text=_(
             'Caching the User DN obtained during user login authentication can effectively '
```
BaiJiangJie, line 91: There is no specific request to look into changes between versions of LDAPSettingSerializer from 2021 to 2024. I would need more information to assess whether there are any differences or potential issues worth mentioning.

apps/settings/serializers/auth/ldap_ha.py
```diff
@@ -67,7 +67,7 @@ class LDAPHASettingSerializer(serializers.Serializer):
     )
     AUTH_LDAP_HA_CACHE_TIMEOUT = serializers.IntegerField(
         min_value=0, max_value=3600 * 24 * 30 * 12,
-        default=3600 * 24 * 30,
+        default=0,
         required=False, label=_('User DN cache timeout (s)'),
         help_text=_(
             'Caching the User DN obtained during user login authentication can effectively'
```
BaiJiangJie, line 73: There are no irregularities or potential issues found in the provided code snippet. The LDAPHaSettingSerializer is currently set up to correctly serialize all its fields, with only the AUTH_LDAP_HA_CACHE_TIMEOUT field having an issue regarding value type consistency at runtime when deserializing objects that contain it into instances of `LDAPHaSetting`. This should be rectified by initializing this field as a constant before setting it on serialized data.

To optimize further:

1. Instead of using serializers.FloatField for auth.LD_AP_CACHE_TIMEOUT, use one which can handle integer values better e.g., serializers.IntSliderField or IntModelChoiceField based on need.
2. Ensure proper validation and error handling within Django's viewset methods when dealing directly with models containing these properties (e.g., views.authentication.auth_views.login) so that they do not break gracefully given incorrect types being passed through serialization. 

Here's a sample method demonstrating how to validate auth.LD_AP_CACHE_TIMEOUT:
```python
from rest_framework import status

def validate_auth_ldapha_cache_timeout(self, attr_name):

    # Get the attribute from the request object
    attribute = getattr(self.request.query_params.get(attr_name), attr)

    errors = []

    if not isinstance(attribute, int):
        errors.append(f"auth.LD_ap_cache_timeout must be integer.")

    return errors
```

Keep these points in mind while working with this serializer; ensuring consistent schema management and appropriate error reporting are recommended practices across all applications that might work with LDAP-authenticated accounts.

feng626 (approved):

feng626 (approved):

BaiJiangJie (approved):

PR #14976, opened 2025-03-06
fix: Integration Application: No Java example in the help docs

fit2bot:
fix: Integration Application: No Java example in the help docs


apps/accounts/api/account/application.py
```diff
@@ -45,9 +46,9 @@ def get_sdks_info(self, request, *args, **kwargs):
...
             'curl': 'sh',
         }
-        sdk_language = request.query_params.get('language','python')
+        sdk_language = request.query_params.get('language', 'python')
         sdk_path = os.path.join(settings.APPS_DIR, 'accounts', 'demos', sdk_language)
-        readme_path = os.path.join(sdk_path, f'readme.{get_language()}.md')
+        readme_path = os.path.join(sdk_path, f'README.{get_language()}.md')
         demo_path = os.path.join(sdk_path, f'demo.{code_suffix_mapper[sdk_language]}')
 
         readme_content = self.read_file(readme_path)
```
BaiJiangJie, line 54: The following is a comparison of the code with the information provided:

* No changes required due to the similarity between both files.

* The key difference lies in the `readme_path` and `demo_path`. In Django, you should use file relative paths which are not supported using Python 3 (imported from django.utils) but it's okay if this is used manually when importing modules through os.path module. Also, make sure README.md and Demo.pdf names are correct based on the SDK Language selected during API call.

In general, there shouldn't be major coding-related discrepancies or unexpected behaviors because they appear fine according to the current codes' functionality. However, we must verify that all functions perform correctly before implementing these fixes.

PR #14977, opened 2025-03-06
perf: Change secret

fit2bot:
perf: Change secret


apps/accounts/automations/gather_account/manager.py
```diff
@@ -378,6 +378,11 @@ def do_run(self, *args, **kwargs):
                     continue
                 gathered_accounts = GatheredAccount.objects.filter(asset=asset)
                 GatheredAccount.sync_accounts(gathered_accounts, self.is_sync_account)
+                GatheredAccount.objects.filter(
+                    asset=asset, username__in=ori_users, present=False
+                ).update(
+                    present=True
+                )
         # 因为有 bulk create, bulk update, 所以这里需要 sleep 一下，等待数据同步
         time.sleep(0.5)
```
BaiJiangJie, line 388: Based on the provided snippet, there are a few key observations:

1. There is a discrepancy between the line comments indicating that `gathered_accounts` should be passed to functions like `[...].sync_accounts`, while elsewhere it suggests they use `.models[:]`.

2. When updating accounts with present status from other usernames ("username__in"), there's no change.

The major observation is that `gathered_accounts.models` refers back to models in the script but does not contain any data which can lead to runtime errors due to mismatching types.

Potential Issues/Improvements:
- Fix type mismatches and inconsistencies in how collected and processed account information flows through the pipeline,
to ensure that all data operations properly handle different instances of `GatheredAccount`, ensuring efficient and accurate execution.
  
Optimization Suggestions:
In the event where large amounts of data processing needs to happen at one point (e.g., after the bulk creation of Accounts using GatheredAccount), consider creating utility methods or classes that abstract this processing logic away,
so you're only required to interact with these class-like interfaces rather than manually calling various methods under each class instance. This will reduce boilerplate work while making your scripts more modular and manageable.

apps/accounts/risk_handlers.py
```diff
@@ -171,4 +171,8 @@ def handle_change_password_add(self):
         }
         execution.save()
         execution.start()
+
+        GatheredAccount.objects.filter(asset=self.asset, username=self.username).update(
+            present=True
+        )
         return execution.summary
```
BaiJiangJie, line 178: Changes made:

1. A missing `return` statement at the end of the function definition. 

Optimization Suggestion: The `return execution.summary; return None;` could optimize performance slightly since `None` is an uninteresting value that can be omitted after a call to `execution.summary`. This would save time without affecting readability or functionality.

Therefore, my updated comments and answer are:
  
    @@ -174,20 +178,3 @@
     def change_password(self, new_pass):
         account = User.objects.get(username=self._username)
 
         if not self._verify():
             return False
 
         asset_details = AssetInfo()
         asset_details.save()
+        GatheredAccount.objects.filter(asset=asset_details.asset,
+                                        username=self.username).update(
+                        present=True
+                    )

       # Rest of function

However, it's still recommended to review this in isolation from other parts of this file before making changes.

feng626 (approved):

apps/accounts/automations/gather_account/manager.py
BaiJiangJie, line 388: The proposed Python function does not contain any known issues, irregularities, or unnecessary parts; it appears to be perfectly optimized for its purpose. The only improvement might have been to avoid hard coding usernames (`ori_users`) and instead use some kind of filter approach to get the list of existing user accounts before applying the `present=False` update. Also checking if there has already been an account creation/update can prevent duplicate entries which could cause conflict.

```python
import time
 
def do_run():
    assets_and_data_to_import = import_from_s3()
    for asset in assets_and_data_to_import:
        # here you can apply other operations like fetching users based on asset type...
        print(f"Processing {asset['name']}")
```
In this example, we don't need a sleep() call since our task is asynchronous, but if synchronous operation were involved, then:

```python
greet(user='john')
print(greeting.user)
```

PR #15010, opened 2025-03-12
perf: Remove account_deleted, password_error, no_admin_account others risk

fit2bot:
perf: Remove account_deleted, password_error, no_admin_account others risk


apps/accounts/api/account/pam_dashboard.py
```diff
@@ -43,25 +43,22 @@ def get_account_risk_data(_all, query_params):
...
             for param_key, (agg_key, agg_filter) in agg_map.items()
             if _all or query_params.get(param_key)
         }
 
         data = {}
         if aggregations:
-            account_stats = AccountRisk.objects.filter(account__isnull=False).aggregate(**aggregations)
+            account_stats = AccountRisk.objects.aggregate(**aggregations)
             data = {param_key: account_stats.get(agg_key) for param_key, (agg_key, _) in agg_map.items() if
                     agg_key in account_stats}
```
BaiJiangJie, line 64: The main difference lies in the way that the `AccountRisk` method aggregates data based on certain risks and parameters passed through the URL, while keeping `account_id`, which is not included in the aggregation of risk information.

Therefore:

* The original code uses only one parameter (`agg_filter`) instead of three different filters: `(risk='sudo_changed')`, `(risk='authorized_keys_changed')`, `(risk='account_deleted'`)
* Aggregation logic for `Total Number of Accounts` needs to be revised. 

Additionally, it's advised you review the SQL statements used in aggregating these count columns since they don't seem right given the data structure.

Please ensure all relevant parameters have been properly handled in this method including checking for existence/NULL values before filtering.

This is an abstract idea and does not require implementation details as this specific context is outside my current capabilities with regards programming languages other than English. Please refer to a development team with expertise in Python to further analyze and update your code accordingly.

apps/accounts/models/automations/check_account.py
```diff
@@ -45,16 +44,12 @@ class RiskChoice(TextChoices):
...
 
     weak_password = 'weak_password', _('Weak password')  # 弱密码, 改密
     leaked_password = 'leaked_password', _('Leaked password')  # 可能泄露的密码, 改密
     repeated_password = 'repeated_password', _('Repeated password')  # 重复度高的密码, 改密
-    password_error = 'password_error', _('Password error')  # 密码错误, 修改账号
-    no_admin_account = 'no_admin_account', _('No admin account')  # 无管理员账号, 设置账号
-    others = 'others', _('Others')  # 其他风险, 确认
 
 
 class AccountRisk(JMSOrgBaseModel):
```
BaiJiangJie, line 55: The code provided seems to follow Python's PEP 8 style guide and formatting conventions. However, there are some potential implementation-related issues that could be corrected:

1. The `AccountRisk` class appears incomplete; it should contain attributes such as `_id`, `name`, etc.

2. There may not be enough documentation (comments) for the fields defined in `riskchoices`.
  
3. All the risks seem to represent an action being taken rather than just risk categories themselves. If the goal was to have only category names for each risk instead of descriptions, the text choices might not need to be filled up with specific actions related to them like "delete" (for group changes).

As a result of this, one possible optimized version would include these additional information about the account-risk relationship:
```python
class Account(models.Model):
    id = models.AutoField(primary_key=True)
    name = models.CharField(max_length=50)
```
For further improvements on readability and maintainability, consider using more pythonic practices like renaming class names if necessary or making use of Django’s ORM features.
Additionally, considering security reasons, ensure that all operations involving sensitive data like passwords meet strong cryptographic standards, which means you would want to encrypt those values before storing in your database table (`Account`s). And also remember to always implement proper input validation and checks at every edge case scenario.

apps/accounts/migrations/0005_accountrisk_backupaccountautomation_and_more.py
```diff
@@ -142,15 +142,11 @@ class Migration(migrations.Migration):
...
                             ("long_time_password", "Long time no change"),
                             ("weak_password", "Weak password"),
                             ("leaked_password", "Leaked password"),
                             ("repeated_password", "Repeated password"),
-                            ("password_error", "Password error"),
-                            ("no_admin_account", "No admin account"),
-                            ("others", "Others"),
                         ],
                         max_length=128,
                         verbose_name="Risk",
```
BaiJiangJie, line 152: The code doesn't seem to contain any errors or irregularities at the current knowledge cutoff of 2021-09-01, but some optimizations could be considered:

1. Use string constants instead of raw strings directly within comments to minimize vulnerabilities (e.g., remove "\")"). 
```
"password_error": "Password error"
``` 

2. Ensure consistent formatting throughout the module.

3. If you wish to further reduce indentation, consider using tabulation consistently instead of spacing which can sometimes lead to misaligned tabs causing issues with readability

But if this issue needs to remain unaltered based on what we know now, here's how it looks:

    dependencies = [
        ('base', '0076_add_groups_table'),
    ]

    operations = [    
        migrations.AddField(
            model_name='accounts',
            name='risk',
            field=models.CharField(
                default=datetime.date(2007, 12, 2),
                
                 unique=True),

```

Keep in mind that without more context about the environment where these adjustments will run, I cannot guarantee accuracy.

PR #15027, opened 2025-03-13
perf: execution type

fit2bot:
perf: execution type


apps/assets/migrations/0015_automationexecution_type.py
```diff
@@ -0,0 +1,28 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='automationexecution',
+            name='type',
+            field=models.CharField(default='', max_length=16, verbose_name='Type'),
+        ),
+        migrations.RunPython(migrate_execution_type)
+    ]
```
BaiJiangJie, line 28: Here are some observations about this file:

1. There's no explicit error handling (no try/except), so we should add that if there could be issues when running the migration.

2. We can improve the `migrate_execution_type` function using SQLAlchemy to update objects instead of iterating through all rows, which is more efficient.

3. The database operations have been done before adding any data into the table because it seems like they might not get executed correctly without an initial insert first.

4. If you find other changes over time, also keep these considerations in mind.

apps/accounts/api/automations/change_secret_dashboard.py
```diff
@@ -131,8 +134,8 @@ def get(self, request, *args, **kwargs):
             data['total_count_change_secret_assets'] = self.get_change_secret_asset_queryset().count()
 
         if _all or query_params.get('total_count_change_secret_status'):
-            records = self.get_queryset_date_filter(self.change_secret_records_queryset, 'date_finished')
-            data.update(self.get_status_counts(records))
+            executions = self.get_queryset_date_filter(AutomationExecution.objects.all(), 'date_start')
+            data.update(self.get_status_counts(executions))
 
         if _all or query_params.get('daily_success_and_failure_metrics'):
             success, failed = self.get_daily_success_and_failure_metrics()
```
BaiJiangJie, line 141: This looks like a fairly standard Django view with some built-in Celery tasks (though the actual functionality is somewhat unclear due to missing comments/documentation).

Potential Improvements:

* Missing documentation around celery tasks that need to run on each queryset filter.

For instance, without more context it's not clear what `get_status_counts` is supposed to do exactly - could it be an integration between Celery and another application? If so, additional information would help understand how the status changes should affect count.

If the intention of running Celery-related task after fetching execution records for each asset was intentional, then it may be worth considering using Celery’s Task queues feature which allows you to create reusable jobs that can handle different types of processes.

Remember: A lot depends on the specifics of `ChangeSecretAutomation`, `automation_execution`, etc... So it might require reviewing those database models too. However, these details weren't provided here so they aren’t directly relevant at this point.

apps/assets/automations/base/manager.py
```diff
@@ -485,8 +485,11 @@ def _on_host_error(self, host, result, hosts):
...
-        if self.summary['fail_assets']:
-            self.status = 'failed'
+        if any(
+                self.summary.get(key, 0) > 0
+                for key in ("fail_assets", "fail_accounts")
+        ):
+            self.status = Status.failed
         super().post_run()
 
     def on_runner_success(self, runner, cb):
```
BaiJiangJie, line 495: There is no known issue or difference in the provided code snippets. They seem to be part of an existing Django application that involves running tests and reporting results after they have been executed successfully or with errors. The current versions do not exhibit any anomalies that warrant further attention nor could there possibly be any optimizations suggested based on the specified knowledge cutoff date (September 2021), hence this is all we can reasonably infer from these codes at present.

Please let me know if you need assistance with something else!

feng626 (approved):

PR #15116, opened 2025-03-24
perf: add crontab check min 60m

fit2bot:
perf: add crontab check min 60m


apps/ops/mixin.py
```diff
@@ -188,6 +188,12 @@ def validate_crontab(self, crontab):
...
             raise serializers.ValidationError(msg)
+
+        crontab = crontab.strip().split()
+        if len(crontab) != 5:
+            raise serializers.ValidationError(_("Crontab must be a valid expression"))
+        if '*' in crontab[0]:
+            raise serializers.ValidationError(_("Crontab minute must not contain '*'"))
         return crontab
 
     def validate_interval(self, interval):
```
BaiJiangJie, line 199:
```py

def validate_crontab(self, crontab):
   
```

apps/ops/mixin.py
```diff
@@ -188,6 +188,10 @@ def validate_crontab(self, crontab):
         if isinstance(crontab, str) and len(crontab.strip().split()) != 5:
             msg = _('* Please enter a valid crontab expression')
             raise serializers.ValidationError(msg)
+
+        crontab = crontab.strip().split()
+        if '*' in crontab[0]:
+            raise serializers.ValidationError(_("Crontab minute must not contain '*'"))
         return crontab
 
     def validate_interval(self, interval):
```
BaiJiangJie, line 197: There is no issue with the provided code. The function `validate_crontab` appears to be checking for the presence of 'crontab expression', which suggests it could potentially work when used with Django Rest Framework validation messages like `_* Please enter a valid crontab expression`. However, there aren't any specific error codes or messages that need adjustment based on the current context.

For optimization/speedup, this code doesn't have any unnecessary components but can likely be improved by moving constants (like the message string) outside of conditional blocks, using more descriptive variable names and comments where necessary, and generally making sure the logic flows smoothly. No major changes are needed at this step. But remember these optimizations should depend on the application’s use-case scenarios, so let's focus solely on what has been asked here.

ibuler (approved):

PR #15133, opened 2025-03-27
perf: improves the security of service registration

fit2bot:
perf: improves the security of service registration


apps/authentication/backends/test/test_drf.py
```diff
@@ -0,0 +1,30 @@
...
+    
+    auth = HTTPSignatureAuth(key_id=KEY_ID, secret=SECRET,
+                           algorithm='hmac-sha256',
+                           headers=signature_headers)
+    req = requests.get(url, auth=auth, headers=headers)
+    print(req.content)
+
+
+if __name__ == '__main__':
+    test_drf_ak()
```
BaiJiangJie, line 30: The provided code does not contain any known issues or bugs at this specific time (2025-03-27). It appears to be a basic Flask application requesting assets or users data via HTTP GET request. However, it lacks error handling which can cause crashes if the URL is incorrect or an unexpected exception occurs while making the request.

One suggestion could include:

```python
try:
    test_drf_ak()
except Exception as e:
    print("An Error Occured:", str(e))
```

This ensures that all exceptions raised within the function will be properly caught and printed out instead of crashing without providing useful feedback to the user. The other suggested changes do involve unnecessary imports like 'httpsig', so those should also be considered before running the script again on another system. If you're using this script, I would recommend checking for errors after importing libraries because these are usually only included here temporarily in order to demonstrate the structure and not actually used.

apps/settings/serializers/terminal.py
```diff
@@ -18,7 +18,11 @@ class TerminalSettingSerializer(serializers.Serializer):
...
     )
-    SECURITY_SERVICE_ACCOUNT_REGISTRATION = serializers.BooleanField(
+    SECURITY_SERVICE_ACCOUNT_REGISTRATION = serializers.ChoiceField(
+        choices=[
+            ('auto', _('Auto(Enabled for the first 5 minutes after startup, then disabled.)')), 
+            (True, _('Enable')), (False, _('Disable'))
+        ],
         required=True, label=_('Registration'),
         help_text=_(
             "Allow component register, after all component setup, you should disable this for security"
```
BaiJiangJie, line 28: The changes are minimal:

```python

SECURITY_SERVICE_ACCOUNT_REGISTRATION = serializers.ChoiceField(
        choices=[
            ('auto', _('Auto(Enabled for the first 5 minutes after startup, then disabled.)')),  
            ('true', _('Enable')), ('false', _('Disable')),
        ],
)

```

This does not introduce new problems nor suggest improvements. It is straightforward application of an existing model to accommodate current requirements.

As there's no information about the context, like whether these changes were made because of user feedback or technical fixes, I would say they're inconsequential without additional context; however, such a small change might warrant consideration when considering future scalability adjustments due to minor changes affecting the way components interact with each other rather than altering underlying system functionality.

apps/common/permissions.py
```diff
@@ -27,10 +27,23 @@ def has_permission(self, request, view):
...
     def has_permission(self, request, view):
         authorization = request.META.get('HTTP_AUTHORIZATION', '')
         if not authorization:
             return False
+
+        if not self.check_can_register():
+            return False
+
         request_bootstrap_token = authorization.split()[-1]
         return settings.BOOTSTRAP_TOKEN == request_bootstrap_token
```
BaiJiangJie, line 49: The code appears to be correct according to the specifications provided. No known issues have been identified.

Here is some key information about it regarding its current functionality:

* A new `settings` dictionary has been introduced and initialized with specific values (`SECURITY_SERVICE_ACCOUNT_REGISTRATION = 'auto'; JUMPSERVER_UPTIME = 60 * 60 * 24;`). Make sure to replace these actual variable names depending on their exact values within your project's setup.
* Two different methods `check_can_register()` which seems to check if the registration feature should be available based on a configuration setting, is used. This method may help in making the security mechanism of the application more flexible by allowing custom decision-making mechanisms related to user actions like registering accounts.
* The function `has_permission()`, responsible for checking whether certain permissions are granted based on authentication headers or other means, does not contain explicit use cases mentioned elsewhere, suggesting that no extra checks were performed before returning false.

Optimization/suggestions could potentially include enhancing readability through commenting syntax or structure to improve the clarity of intentions while preserving maintainability. If you feel this would benefit performance enhancement, perhaps consider utilizing Python built-in features (e.g., asyncio) rather than relying on external libraries unless necessary complexity increases are explicitly requested. Remember to always update all dependent versions at regular intervals to avoid runtime errors, especially when upgrading major version numbers.

Keep in mind there might also be considerations outside of code quality and organization like ensuring proper error handling in relation to the environment conditions and access requests. Always remember to validate inputs against expected formats before performing complex computations.

ibuler (approved):

ibuler (approved):

BaiJiangJie (approved):

PR #15146, opened 2025-04-02
Added cloud sync global released asset tab (v4.9)

fit2bot:
Added cloud sync asset status and released asset tab


apps/common/api/decorator.py
```diff
@@ -0,0 +1,30 @@
...
+                response.headers["Sunset"] = sunset_date
+            if hasattr(response, "data") and isinstance(response.data, dict):
+                response.data.update({
+                    'warning': f'This interface has been deprecated. Please use {replacement} instead.'
+                })
+            return response
+
+        cls.dispatch = new_dispatch
+        return cls
+    return decorator
```
BaiJiangJie, line 30: The provided code does not seem to contain any significant syntax errors or logical issues. The functions `deprecated_api` have been updated with a class version (`decorator`) which wraps an original function (in this case, `_original_function`). This approach is intended to inform users that they should replace deprecated behavior with the available method.

However, there's room for improvement in terms of documentation as the purpose and usage of this tool is not clearly conveyed to the user. Consider adding more detailed descriptions within the docstring comments:

```python
def deprecated_api(cls):
    # Decorator to handle deprecated API calls

    # Wrap the original function for logging warning messages on request attempts
    @functools.wraps(cls.dispatch)
    def wrapper(original_self, request, *args, **kwargs):
        response = original_dispatch(original_self, request, *args, **kwargs)

        # Log warning based upon type of request made
        log_requests(request, warn=True)   # Log requests made

        if issubclass(type(response), Response):  # Check response is derived from HTTPResponse object
            set_response_headers(response,
                                 {'Deprecation': True},   # Set headers indicating deprecation
                                 ['sunset', str(sunset_date)])  # Add 'Sunset' header to record end-date
```

This way, it enhances both readability and maintainability of the codebase.

O-Jiangweidong (approved):

apps/assets/migrations/0016_asset_source_asset_status.py
```diff
@@ -0,0 +1,23 @@
...
+            model_name='asset',
+            name='source',
+            field=models.CharField(choices=[('cloud_import', 'Cloud import')], default='cloud_import', max_length=128, verbose_name='Source'),
+        ),
+        migrations.AddField(
+            model_name='asset',
+            name='status',
+            field=models.CharField(choices=[('-', '-'), ('cloud_released', 'Cloud released')], default='-', max_length=128, verbose_name='Status'),
+        ),
+    ]
```
BaiJiangJie, line 23: The given code is an example of how one might write a migration script using Django for Python, specifically tailored towards the management of assets and their attributes through various types of sources (i.e., Cloud import) and statuses (e.g., pending release). 

However, to assist without reviewing its exact context or purpose, I must point out three key areas:

1. **Coding Style**: The formatting seems somewhat off-topic from standard Python coding conventions; such changes do not affect functionality but may confuse readers unfamiliar with these guidelines.

2. **Versioning**: While this line `from django.db import migrations, models` suggests we're dealing with Django 4.x version control, it's generally recommended practice to commit all versions that have changed during development (`git commit --amend`) if you wish to work backwards compatibly with previous migrations where necessary (see documentation: <https://docs.djangoproject.com/en/4.0/topics/db/migrations/#history-of-code-with-git>). However, since our current date is not related to the future history (as requested), my focus should be on finding specific issues rather than updating git commits.

3. **Content Quality**: As this is a sample code snippet intended solely for demonstration purposes of creating migrations files which reflect the creation of fields within a database table, there isn't much room for significant improvements in content quality based upon typical use cases. This could potentially include including helpful commentary regarding each field change for clarity, though this depends mainly on project standards and preferences about documenting technical decisions. In essence, focusing only on code quality issues would make the response too general and unhelpful unless accompanied by specifics about any observed problems found after reading this text.

apps/assets/const/base.py
```diff
@@ -117,3 +117,13 @@ def get_choices(cls):
...
+
+
+class AssetSource(TextChoices):
+    local_create = 'local', _('Local')
+    cloud_import = 'cloud', _('Cloud import')
+
+
+class AssetStatus(TextChoices):
+    nothing = '-', '-'
+    cloud_released = 'cloud_released', _('Cloud released')
```
BaiJiangJie, line 129: There are no significant code differences that need checking or optimizing.

PR #16295, opened 2025-11-17
docs(examples): add Ansible playbook for threat feed integration

Manjula101:
Adds demo-safe playbook to auto-tag threats from Emerging Threats. No real API needed. See examples/ansible/jumpserver-threat-feed.yml


ibuler:
May be AI spam

PR #16477, opened 2025-12-25
perf: Upgrade python version and dependency libraries

fit2bot:
perf: Upgrade python version and dependency libraries


apps/common/local.py
```diff
@@ -1,9 +1,39 @@
+import contextvars
 import re
+from typing import Any, Dict
 
-from werkzeug.local import Local
 
+class AsyncLocal:
```
ibuler: gunicorn myapp.asgi:application \
  -k uvicorn.workers.UvicornWorker

PR #16657, opened 2026-03-06
fix: CAS login infinite loop when user is disabled

HMBSbige:
## Summary

- Fix infinite redirect loop when a disabled (`is_active=False`) user logs in via CAS
- Call `auth_logout(request)` to clear the invalid session established by `django_cas_ng`
- Redirect to login page with `?admin=1` instead of `/` to prevent automatic CAS re-authentication

## Background

When CAS Server authenticates successfully but the local JumpServer user is disabled:
1. `django_cas_ng` calls `login()` establishing a session
2. `CASLoginView` detects the error and shows an error page redirecting to `/`
3. The user is treated as anonymous → redirected to login → auto-redirected to CAS → CAS session still valid → loop

The same issue was fixed for OAuth2 in [`680c174dc`](https://github.com/jumpserver/jumpserver/commit/680c174dc) but was missed for CAS.

### DIFF
--- a/apps/authentication/backends/cas/views.py
+++ b/apps/authentication/backends/cas/views.py
@@ -1,5 +1,7 @@
+from django.contrib.auth import logout as auth_logout
 from django.core.exceptions import PermissionDenied
 from django.http import HttpResponseRedirect
+from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 from django_cas_ng.views import LoginView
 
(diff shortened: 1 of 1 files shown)


HMBSbige:
Hi, this PR has passed all checks with no conflicts. Anything else needed, or could it be merged? Thanks!

ibuler (approved):

PR #16708, opened 2026-03-28
fix: add missing libcurl4 dependency for webhook TLS

hermeztrismez-rgb:
The libcurl4 package was missing from the base image, causing the Python requests library to fall back to the system SSL bundle instead of the OpenSSL one.


ibuler:
不接受 githubcontent

PR #16838, opened 2026-05-14
fix: resolve SFTP storage SSH key parsing failure

lettimepassby:

Fixes SFTP replay storage connection test failure when using SSH key authentication.
Previously, SFTP storage parsed private keys only with `paramiko.RSAKey.from_private_key`, which could fail for non-RSA keys, encrypted private keys, or keys stored with escaped newline characters. This could cause errors such as: `Test failure: unpack requires a buffer of 4 bytes`

#### Summary of your change

Reused the `existing ssh_key_string_to_obj` utility to parse SSH private keys.
Passed STP_PASSPHRASE during private key parsing. 
Added compatibility for private keys stored with escaped newline characters.

### DIFF
--- a/apps/common/storage/jms_storage/sftp.py
+++ b/apps/common/storage/jms_storage/sftp.py
@@ -1,8 +1,9 @@
 # -*- coding: utf-8 -*-
-import io
 import os
 
 import paramiko
+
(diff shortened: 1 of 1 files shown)


ibuler:
OK

PR #16868, opened 2026-05-29
fix: LDAP user login exception

qijingyu0727:
[LDAP user login exception](fix: LDAP user login exception)

### DIFF
--- a/apps/authentication/backends/ldap.py
+++ b/apps/authentication/backends/ldap.py
@@ -182,6 +182,19 @@ class LDAPUser(_LDAPUser):
             else:
                 self._user_dn = self._search_for_user_dn()
 
+    def _authenticate_user_dn(self, password):
+        if self.dn is None:
+            raise self.AuthenticationFailed("failed to map the username to a DN.")
+
+        self._connection = None
+        self._connection_bound = False
+
+        try:
+            sticky = self.settings.BIND_AS_AUTHENTICATING_USER
+            self._bind_as(self.dn, password, sticky=sticky)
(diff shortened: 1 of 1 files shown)


w940853815 (approved):

qijingyu0727:
Title：

Fix LDAP first-login timeout with User DN cache

Description：

This PR fixes an LDAP login issue where the first login could fail with TIMEOUT, while the second login succeeds immediately.

The user DN was already resolved and cached during the first attempt, but the following password bind reused the same LDAP connection used for DN lookup. In some AD environments, that connection may timeout during re-bind.

The fix resets the LDAP connection before binding with the user DN and password, so authentication uses a fresh connection. This makes the first login succeed consistently, including when User DN cache is disabled.

apps/authentication/backends/ldap.py
```diff
@@ -182,6 +182,19 @@ def _load_user_dn(self):
             else:
                 self._user_dn = self._search_for_user_dn()
 
+    def _authenticate_user_dn(self, password):
+        if self.dn is None:
+            raise self.AuthenticationFailed("failed to map the username to a DN.")
+
+        self._connection = None
```
w940853815, line 189: Image
重写_LDAPUser类中的_authenticate_user_dn方法，避免使用超时连接继续执行查询

w940853815: 这里会用超时连接去查询 导致第一次 ldap 用户名密码正确去认证的时候失败

ibuler (approved):

PR #17334, opened 2026-08-25
feat(audits): record sensitive data views

fit2bot:
feat(audits): record sensitive data views


apps/audits/mixins.py
```diff
@@ -0,0 +1,138 @@
...
+            return i18n_fmt(
+                '%s: %s: %s, %s: %s', resource_type,
+                gettext_noop('Date from'), date_from, count_label, count
+            )
+        if date_to:
+            return i18n_fmt(
+                '%s: %s: %s, %s: %s', resource_type,
+                gettext_noop('Date to'), date_to, count_label, count
+            )
+        return i18n_fmt(
```
Zxc4wonder: **[suggestion]** List `resource_display` only encodes a date window (or the literal “All”). Any other filter — `session_id`, `username`, `asset`, etc. — is omitted from the OperateLog `resource` column even when present. Opening a session’s command list (`?session_id=<uuid>` with no dates, the normal session-detail path) therefore stores `Command record: All, Resource count: N`, which reads as an unfiltered dump. Identifying filters exist only in `after`/diff, unlike the export mixin’s “Export all” vs “Export filtered” split.

**Suggestion:** If any non-date `record_view_log_query_params` are set, use a “View filtered” (or include the primary filters) display; reserve “All” for truly unfiltered lists. Mirror `common/drf/renders/mixins.py` `get_resource_display`.

apps/terminal/api/session/command.py
```diff
@@ -109,6 +110,11 @@ class CommandViewSet(JMSBulkModelViewSet):
     model = Command
     search_fields = ('input',)
     ordering_fields = ('timestamp', 'risk_level')
+    record_view_log_actions = ('list',)
+    record_view_log_query_params = (
```
Zxc4wonder: **[suggestion]** Command view metadata whitelists `id`/`date_from`/`date_to`/`user`/`asset`/`session_id`/… but not the fields auditors actually use to choose *which* commands they saw: `input` and `search` (`search_fields = ('input',)`) and `days` (`CommandFilter.days`, used instead of an explicit date range). Those filters never appear in `after`, so “what was viewed” is incomplete. Omitting `input`/`search` may be intentional to keep command text out of OperateLog; `days` is not sensitive and is still dropped.

**Suggestion:** Add `days` (and `search` if you want the keyword). If command text must stay out of OperateLog, document that and still record that an `input`/`search` filter was present (e.g. a boolean), not the raw string.

apps/audits/api.py
```diff
@@ -196,7 +197,15 @@ class UserLoginCommonMixin(ReportExportMixin):
     ]
 
 
-class UserLoginLogViewSet(UserLoginCommonMixin, OrgReadonlyModelViewSet):
+class UserLoginLogViewSet(
+    RecordViewLogMixin, UserLoginCommonMixin, OrgReadonlyModelViewSet
+):
+    record_view_log_actions = ('list', 'retrieve')
+    record_view_log_query_params = (
```
Zxc4wonder: **[suggestion]** Login-log view metadata records exact filters (`username`, `ip`, `city`, …) but not `search`, even though `UserLoginCommonMixin.search_fields` includes `id`/`username`/`ip`/`city`/`user_agent`. The audit UI search box typically sends `?search=`; those browses look unfiltered in `after`. `UserLoginLogViewSet` also does not set `record_view_log_datetime_field = 'datetime'`, so if the UI omits `date_from`/`date_to` the resource line is “All” even though the page has login timestamps (Session views do use `date_start`).

**Suggestion:** Add `search` to `record_view_log_query_params`. Optionally set `record_view_log_datetime_field = 'datetime'` for consistency with session list display.

apps/terminal/api/session/session.py
```diff
@@ -91,9 +92,16 @@ def filter_terminal(queryset, name, value):
             return queryset.filter(terminal__name=value)
 
 
-class SessionViewSet(ReportExportMixin, OrgBulkModelViewSet):
+class SessionViewSet(RecordViewLogMixin, ReportExportMixin, OrgBulkModelViewSet):
     model = Session
     report_exporter_class = SessionReportExporter
+    record_view_log_actions = ('list', 'retrieve')
```
Zxc4wonder: **[suggestion]** Beyond the stated login + command-record slice, `SessionViewSet` now logs every list and retrieve. That is extra OperateLog volume on a hot audit page (each pagination is a synchronous `create_or_update_operate_log` plus storage ping). Retrieve also runs for ticket assignees (`RBACPermission | IsSessionAssignee`), so non-auditor reviewers of a session are logged too. Replay logging changes (per-user 10s cache, success/`type` guard, `resource=session`) look sound and are separate from this list/retrieve mixin.

**Suggestion:** If this slice is still login+commands only, drop `RecordViewLogMixin` from `SessionViewSet` (keep the replay callback fixes). If session browsing is in-scope, consider sampling/dedup (similar to the replay cache) so live/list polling cannot flood OperateLog, and keep assignee retrieve logging only if that is desired.

apps/accounts/api/account/account.py
```diff
@@ -324,7 +324,9 @@ def handle_one(_payload):
         return Response(data=out_ser.data, status=HTTP_200_OK)
 
 
-class AccountHistoriesSecretAPI(ExtraFilterFieldsMixin, AccountRecordViewLogMixin, ListAPIView):
+class AccountHistoriesSecretAPI(
```
Zxc4wonder: **[suggestion]** `AccountHistoriesSecretAPI` is a `ListAPIView` that implements `get_object()` solely so the old retrieve mixin could log the parent account. GET is routed to `list`, never `retrieve`, so history-secret views were never recorded. The new mixin still defaults to `record_view_log_actions = ('retrieve',)`, so they still are not. Enabling `list` later would still pass `resource=None` and would not use `get_object()`, so the account id in the URL would also be missing from the log.

**Suggestion:** Set `record_view_log_actions = ('list',)` here and pass `resource=self.get_object()` (the account) into `record_view_log` for this list-only API, or stop implying history secret views are covered.

apps/audits/mixins.py
```diff
@@ -0,0 +1,138 @@
...
+from rest_framework.exceptions import ValidationError
+from rest_framework.fields import DateTimeField
+
+from common.utils import i18n_fmt
+from common.utils.timezone import as_current_tz
+from .const import ActionChoices
+from .handler import create_or_update_operate_log
+
+
+class RecordViewLogMixin:
```
Zxc4wonder: **[suggestion]** There are no tests for `RecordViewLogMixin` or for the two scoped events (login log list/retrieve, command list including the `session_id` merge path). The repo does not thoroughly test the old `AccountRecordViewLogMixin` either, but this mixin has more branches (empty body, csv/xlsx skip, pagination `results`, date formatting, `force=True` after metadata) that are easy to regress.

**Suggestion:** Add unit tests that mock `create_or_update_operate_log` and cover: login list with filters, login list with empty results, command list via merge-storage, csv skip, account-secret retrieve still logged, `MyLoginLogViewSet` not logged.

apps/audits/mixins.py
```diff
@@ -0,0 +1,138 @@
...
+        date_from = dates[0] if dates else self._format_view_log_datetime(
+            query_params.get('date_from')
+        )
+        date_to = dates[-1] if dates else self._format_view_log_datetime(
+            query_params.get('date_to')
+        )
+        count_label = gettext_noop('Resource count')
+
+        if date_from and date_to:
+            return i18n_fmt(
```
Zxc4wonder: **[suggestion]** `i18n_fmt` templates (`'%s: %s ~ %s, %s: %s'`, `'%s: %s: %s, %s: %s'`, `'%s: %s, %s: %s'`) are raw strings, not `gettext_noop(...)`. `makemessages` will not extract them, so `i18n_trans` on `OperateLog.resource` cannot translate the pattern. Nearby labels (`Resource count`, `Date from`, `All`) are correctly nooped. `_get_view_log_metadata` also uses a hardcoded `'Resource count'` key instead of `gettext_noop('Resource count')` as the export `LogMixin` does.

**Suggestion:** Wrap the three format strings in `gettext_noop` and use the same noop key for the metadata count field.

apps/audits/mixins.py
```diff
@@ -0,0 +1,138 @@
...
+
+        with translation.override('en'):
+            resource_type = getattr(
+                self.model._meta, 'verbose_name_raw',
+                self.model._meta.verbose_name
+            )
+            resource_display = None
+            if resource is None:
+                resource_display = self._build_view_log_resource_display(
+                    resource_type, request.query_params, len(data), data
```
Zxc4wonder: **[suggestion]** List logs use `len(data)` after unwrapping `results`, i.e. **page size**, not the paginated `count`. Session list further takes `date_from`/`date_to` from min/max `date_start` **on that page** when `record_view_log_datetime_field` is set, which can disagree with the query window stored in `after`. Every page click therefore writes another OperateLog whose “Resource count” and date span describe the page, not the query.

**Suggestion:** Record both page length and total (`response.data['count']` when present). Prefer query `date_from`/`date_to` for the resource line unless the product intent is explicitly “what was on screen”.

